PR #1, opened 2015-10-11
FREEMARKER-1 OverrideResponseContentType init param in FreemarkerServlet

woonsan:
Added OverrideResponseContentType init param, setting it to true, for backward compatibility.
Also, added a unit test with spring-test dependency (for easier servlet api mocking).
Please take a review. (Once this accepted, FREEMARKER-2 should be easier because of the unit test.)


ddekany:
Can you please re-base this on the 2.3-gae branch? See http://freemarker.org/sourcecode.html

woonsan:
New PR made on 2.3-gae branch: https://github.com/apache/incubator-freemarker/pull/5
Will close this one. Thanks!

PR #3, opened 2015-10-11
update gitignore file to ignore Intellij project files and MacOS and Ubuntu temp files

cnapagoda:


ddekany:
It's certainly sounds stupid to require it for a change like this, but do you have an ICLA at ASF? Otherwise I can't pull... that's the rule.

ddekany:
And actually, it must be based on the 2.3-gae branch, not the master...

PR #8, opened 2015-11-03
FREEMARKER-7 Build with Maven

woonsan:
Initial PR to build with Maven instead of Ant/Ivy:
- Adding initial pom.xml with javacc plugin configuration, dependencies, etc.
- To support multiple incompatible jython versions, I decided to use Reflection APIs for old jython versions because those two classes for the older version of jython APIs are very small and it seems more effective to use Reflection API for those than using complex dependencies and separate compilations.
- To support older version of jsp-api, I simply added an unsupported operation, by which both versions can work without any problem at runtime, while supporting multiple provided jsp-api versions at runtime. I think this approach is simpler and more effective.
- Still I noticed that two unit test operations failing (RealServletContainertTest#basicTrivial() and RealServletContainertTest#basicELFunctions()). I think we'd better separate modules for this kind of real web application runtime tests. Until that, I decided to mark those two operations as @Ignore, to also ask for other's opinions.

Anyway, you can run the following command. Then it will create a freemarker-x.x.x.jar in target/ dir:

```
    mvn clean package
```


ddekany:
For the record, this was discussed elsewhere with Woonsan back then. As things stand today, we will probably go for Gradle for its greater flexibility. Someone said recently that he tries to do that. So we will see if we like Gradle better.

woonsan:
Closing this as this wasn't completed and we have gradle build now.

PR #10, opened 2016-03-17
How to get httprequest in TemplateDirectiveModel

cxc3214:
my code :

``` java
public class ComboTag implements TemplateDirectiveModel{
    @Override
    public void execute(Environment env, Map params, TemplateModel[] loopVars, TemplateDirectiveBody body) throws TemplateException, IOException {
//I want to  get the httprequest  here.
        Writer out = env.getOut();
    }
```


ddekany:
I guess it's a question, not a pull request. FreeMarker is not bound to Servlets, so exposing such objects and how that's done is up to the embedding framework. For example, you could attach the `HttpRequest` to the `Environment` with `setCustomAttribute` before starting the template processing, then get it with `getCustomAttribute`.

PR #12, opened 2016-05-25
Add generic type arguments to "params" in TemplateDirectiveModel

yankee42:
The API for writing own directives could be clearer with generics for the `params`-Parameter. Since the project requires java language level 1.5 and due to the compile-time only nature of generics I suppose that just nobody added them yet, so here is a start.

### DIFF
--- a/src/main/java/freemarker/core/Environment.java
+++ b/src/main/java/freemarker/core/Environment.java
@@ -391,13 +391,13 @@ public final class Environment extends Configurable {
      */
     @Deprecated
     public void visit(final TemplateElement element,
-            TemplateDirectiveModel directiveModel, Map args,
+            TemplateDirectiveModel directiveModel, Map<String, TemplateModel> args,
             final List bodyParameterNames) throws TemplateException, IOException {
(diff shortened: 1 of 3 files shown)


ddekany:
That's deliberate actually. Unfortunately, adding generic parameters to public API-s isn't backward compatible (and these API-s predate Java 5). When you update the FreeMarker dependency, and you recompile some dependent code, it may breaks with compilation errors. And yes, if you only have 3rd party jar-s that call FreeMarker, you won't get any errors, but it's maybe still too much of compatibility breaking for the 2.3.x line.

yankee42:
Thanks for looking into it.  I have to admit I could not quite follow when compilation errors might occur. To not die dump, could you enlighten me with an example?

ddekany:
People already call these old API-s with types that have type parameters. That works as Java is lenient when it has to "convert" between a type with type parameters and the raw form of the same type. But if the types in the API will have type parameters, the compiler will be picky. Like someone might uses a `Map<String, TemplateNumberModel>`, and that's not compatible with the proposed `Map<String, TemplateModel>`. (BTW, that should be `Map<String, ? extends TemplateModel>`.)

PR #13, opened 2016-08-19
Minor fixes to Eclipse setup notes after the update to JDK 8.

jacopoc:
There is a chance that the following sentence should be updated:
"Compiler Compliance Level" to "1.5"

### DIFF
--- a/README
+++ b/README
@@ -152,7 +152,7 @@ apply it to your development environment:
     This profile uses space-only indentation policy and 120 character line
     width, and formatting rules that are pretty much standard in modern Java.
   - Java -> Installed JRE-s:
-    Ensure that you have JDK 6 installed, and that it was added to Eclipse.
+    Ensure that you have JDK 8 installed, and that it was added to Eclipse.
     Note that it's not JRE, but JDK.
 - Create new "Java Project" in Eclipse:
   - In the first window popping up:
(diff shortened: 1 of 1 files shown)


ddekany:
The usage of 1.6 (which you have changed) is not an oversight, nor is the 1.5 compiler compliance level. At the moment we promise Java 1.5 compatibility. However, we also use some (or maybe just one...) 1.6 API-s if they are available. Because the API differences between 1.5 and 1.6 are minor, the least painful is using 1.6 and then the build will still catch if you link to a non-1.5 API. So that's why it's like it is.

jacopoc:
Thank you Daniel for the explanation.
I was actually confused by the following sentence from the same README file:
"You need JDK 8(!), Apache Ant and Ivy to be installed."

ddekany:
Java 8 is because of the build tools. Some build tools needs Java 7 I believe, so you need one more JDK over 5 and 6, and then why not Java 8, rather than 7. Docgen actually requires 8 for a while.

jacopoc:
Thank you Daniel.
So, if I am getting it right, the dependency on JDK 8 (or 7) is only for the creation of the jar, not for its usage. And the Eclipse setup steps are about setting up how the jar is built, not about the JDK used in the build process (here I am guessing since I am not an Eclipse user).
If this is the case I will close/discard this branch.

ddekany:
The Java version chosen in Eclipse is a compromise made because you can't specify different Java version for different java files inside the same project. The jar is built by Ant, and has nothing to do with Eclipse (or ant other IDE). Anyway, yes, this can be closed as I see no problems.

PR #15, opened 2016-09-21
Configure Junit Tests Localization. 

mauronunez:
This allows testing in machines with a different locale ( mine is es_cl )

### DIFF
--- a/src/test/java/freemarker/core/ObjectBuilderSettingsTest.java
+++ b/src/test/java/freemarker/core/ObjectBuilderSettingsTest.java
@@ -48,7 +48,7 @@ import freemarker.cache.MruCacheStorage;
 import freemarker.cache.TemplateLoader;
 import freemarker.core.subpkg.PublicWithMixedConstructors;
 import freemarker.ext.beans.BeansWrapper;
-import freemarker.ext.jython.JythonWrapper;
+
 import freemarker.template.Configuration;
 import freemarker.template.DefaultObjectWrapper;
 import freemarker.template.ObjectWrapper;
@@ -382,13 +382,7 @@ public class ObjectBuilderSettingsTest {
         assertFalse(bw.isExposeFields());
     }
 
-    @Test
-    public void jythonWrapperTest() throws Exception {
-        JythonWrapper jw = (JythonWrapper) _ObjectBuilderSettingEvaluator.eval(
-                "freemarker.ext.jython.JythonWrapper()",
-                ObjectWrapper.class, false, _SettingEvaluationEnvironment.getCurrent());
-        assertSame(JythonWrapper.INSTANCE, jw);
-    }
+
 
     @Test
     public void configurationPropertiesTest() throws TemplateException {
--- a/src/test/java/freemarker/core/OutputFormatTest.java
+++ b/src/test/java/freemarker/core/OutputFormatTest.java
@@ -322,14 +322,14 @@ public class OutputFormatTest extends TemplateTest {
     public void testNumericalInterpolation() throws IOException, TemplateException {
         getConfiguration().setRegisteredCustomOutputFormats(Collections.singleton(DummyOutputFormat.INSTANCE));
         assertOutput(
-                "<#ftl outputFormat='dummy'>#{1.5}; #{1.5; m3}; ${'a.b'}",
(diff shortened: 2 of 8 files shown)


src/test/java/freemarker/core/ObjectBuilderSettingsTest.java
```diff
@@ -382,13 +382,7 @@ public void defaultObjectWrapperTest() throws Exception {
         assertFalse(bw.isExposeFields());
     }
 
-    @Test
```
ddekany: I suppose this was deleted accidentally.

src/test/java/freemarker/core/OutputFormatTest.java
```diff
@@ -322,14 +322,14 @@ public void testAutoEscapingSettingLayers() throws Exception {
     public void testNumericalInterpolation() throws IOException, TemplateException {
         getConfiguration().setRegisteredCustomOutputFormats(Collections.singleton(DummyOutputFormat.INSTANCE));
         assertOutput(
-                "<#ftl outputFormat='dummy'>#{1.5}; #{1.5; m3}; ${'a.b'}",
+                "<#ftl outputFormat='dummy'><#setting locale='en_US'>#{1.5}; #{1.5; m3}; ${'a.b'}",
```
ddekany: That's a good catch, but the locale should be simply set once in the `Configuration`, rather that in each FTL snippet again and again. In fact, the `TemplateTest` should do that automatically, as it's pointless to run tests with a random locale.

src/test/java/freemarker/manual/CustomFormatsExample.java
```diff
@@ -49,7 +50,9 @@ public void aliases1() throws IOException, TemplateException {
 
         Map<String, TemplateDateFormatFactory> customDateFormats
                 = new HashMap<String, TemplateDateFormatFactory>();
-        customDateFormats.put("fileDate", new AliasTemplateDateFormatFactory("dd/MMM/yy hh:mm a"));
+        Map<Locale, String> m=new HashMap<Locale,String>();
```
ddekany: This change is should be unnecessary, because if you don't specify a locale `Map`, the specified default pattern will be used regardless of your locale. Does this test fail for you without the `Map`? That can be a symptom of some other bug then.

src/test/resources/freemarker/manual/CustomFormatsExample-alias1.ftlh
```diff
@@ -16,6 +16,8 @@
   specific language governing permissions and limitations
   under the License.
 -->
+<#setting locale="en_us">
```
ddekany: Indeed, but then the `locale` and `time_zone` should be set in the `ExamplesTest` class, rather that in each template. Especially because otherwise the examples can't be copy-paster from/to the Manual.

ddekany:
Do you plan to work in this more?

mauronunez:
Yes, I want to continue with this. Just give some time to learn this github
interface :-)

2017-02-27 3:25 GMT-03:00 ddekany <notifications@github.com>:

ddekany:
Sorry, we are close to release voting, so I took this as a bug report, and fixed the issues independently:
https://github.com/apache/incubator-freemarker/commit/738cef854c5341fe75243881fd454577151064d2
https://github.com/apache/incubator-freemarker/commit/649d85d180cdd3df87a79e2357d9d77a384dcccf

Thanks for pointing out these problems anyway!  Also, it would be helpful if you can run the tests with the current version, to see if it still contains unwanted assumptions about the test environment.

mauronunez:
Great!

I want to be able to cooperate soon. I will check with my locale the new
versión.

Regards,

Mauricio

El mar. 14, 2017 5:52 PM, "ddekany" <notifications@github.com> escribió:

Sorry, we are close to release voting, so I took this as a bug report, and
fixed the issues independently:
738cef8
<https://github.com/apache/incubator-freemarker/commit/738cef854c5341fe75243881fd454577151064d2>
649d85d
<https://github.com/apache/incubator-freemarker/commit/649d85d180cdd3df87a79e2357d9d77a384dcccf>

Thanks for pointing out these problems anyway! Also, it would be helpful if
you can run the tests with the current version, to see if it still contains
unwanted assumptions about the test environment.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/apache/incubator-freemarker/pull/15#issuecomment-286556189>,
or mute the thread

PR #16, opened 2016-12-27
Add OSGI MANIFEST.MF during ide-dependencies

chrisrueger:
Hi Daniel,
as discussed in https://www.mail-archive.com/dev@freemarker.incubator.apache.org/msg00687.html I finally found time to commit the changes. 

Basically what the 4 commits are doing:

1. during build of target ide-dependencies the META-INF folder inside generated freemarker.jar is extracted
2. for that I adjusted the build.xml and README to point out this change. The ide-dependencies target now requires the freemarker.jar to be present. before it was not needed. 

**Why is this useful?** 

Because I need the META-INF/MANIFEST.MF in my Eclipse / OSGI at **development time** so that other bundles in the same Eclipse workspace  recognize my Freemarker Workspace project (which I imported before based on the README description).

In our company we have a Eclipse / OSGI product and we would like to include the Freemarker source code project directly in eclipse. This makes it much easier to contribute to Freemarker, because we can directly test it during our own development. 

Best regards,
Christoph

### DIFF
--- a/README
+++ b/README
@@ -139,7 +139,7 @@ apply it to your development environment:
 
 - Install Ant and Ivy, if you haven't yet; see earlier.
 - From the command line, run `ant clean javacc ide-dependencies`. (Note that
-  now the "ide-dependencies" and  "build/generated-sources" was created.)
+  now the folders "ide-dependencies", "build/generated-sources" and "META-INF" were created.)
 - Start Eclipse
 - You may prefer to start a new workspace (File -> "Switch workspace"), but
   it's optional.
--- a/README
+++ b/README
@@ -138,7 +138,7 @@ different version or an entierly different IDE, still read this, and try to
 apply it to your development environment:
 
 - Install Ant and Ivy, if you haven't yet; see earlier.
-- From the command line, run `ant clean javacc ide-dependencies`. (Note that
+- From the command line, run `ant clean jar javacc ide-dependencies`. (Note that
   now the folders "ide-dependencies", "build/generated-sources" and "META-INF" were created.)
 - Start Eclipse
 - You may prefer to start a new workspace (File -> "Switch workspace"), but
--- a/README
+++ b/README
@@ -138,7 +138,7 @@ different version or an entierly different IDE, still read this, and try to
 apply it to your development environment:
 
 - Install Ant and Ivy, if you haven't yet; see earlier.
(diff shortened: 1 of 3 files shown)


ddekany:
Thanks, I will look into this when I get to it. It seems that you don't have a CLA at Apache; if that's right, it must be addressed before I can merge, so please see http://www.apache.org/dev/new-committers-guide.html#cla.

README
```diff
@@ -138,7 +138,7 @@ different version or an entierly different IDE, still read this, and try to
 apply it to your development environment:
 
 - Install Ant and Ivy, if you haven't yet; see earlier.
-- From the command line, run `ant clean javacc ide-dependencies`. (Note that
+- From the command line, run `ant clean jar javacc ide-dependencies`. (Note that
```
ddekany: `jar` implies `javacc`, so `javacc` can be omitted.

ddekany:
I wonder if the extracted `META-INF` can be (and should be) put inside `ide-dependencies` (or inside a `build` subdirectory). Also, even if `META-INF` is in the root as now, don't you need some additional step for Eclipse to recognize it?

chrisrueger:
> I wonder if the extracted META-INF can be (and should be) put inside ide-dependencies (or inside a build subdirectory). Also, even if META-INF is in the root as now, don't you need some additional step for Eclipse to recognize it?

Actually for the Eclipse PDE tools the META-INF/MANIFEST.MF needs to be on the project root. That is the case for all the OSGI bundles I know. I cannot prove it with an official reference, but I found the following Eclipse bug: https://bugs.eclipse.org/bugs/show_bug.cgi?id=153023 and https://bugs.eclipse.org/bugs/show_bug.cgi?id=154079#c11 
Here they discuss the option to change the location of the MANIFEST.MF file and why it is where it is. This is still open and seems to be a huge change.

README
chrisrueger: thanks for pointing out. pushed the change.

ddekany:
OK, then that `META-INF` directory will be good where it is now. When Apache receives your CLA I will merge this.

chrisrueger:
Apache CLA is submitted. Waiting for approval.

ddekany:
ASF have confirmend receiving the CLA. Thank you, merged!

PR #18, opened 2017-04-04
Minor fixes in the manual

pyxide:
Hi,
the title says it all.
Olivier

### DIFF
--- a/src/manual/en_US/book.xml
+++ b/src/manual/en_US/book.xml
@@ -1012,7 +1012,7 @@ All Rights Reserved.
  |   |
  |   +- (1st) = 20
  |   |
- |   +- (2st) = 14
+ |   +- (2nd) = 14
  |   |
  |   +- (3rd) = 42
  |   |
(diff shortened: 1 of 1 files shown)


ddekany:
Hi, thanks for the fixes! However, I can only merge them if you have a signed Committer License Agreement at the Apache Software Foundation. You can submit one in e-mail; see here: http://www.apache.org/dev/new-committers-guide.html#cla

pyxide:
I've sent the signed agreement by email. It was much more work than my pull request, I had to rip some code from stackoverflow to compose the PDF document with the scans!
I'll keep you posted.

ddekany:
Apache has received your CLA, so I will be able to merge this now (and BTW now you can contribute to any other Apache projects as well). As of the PDF issues, you are supposed to print the thing as is, fill it by hand, scan the whole page, and save the resulting single big image into a PDF. So no special software is needed.

pyxide:
Thank you for the follow-up and the information.

PR #20, opened 2017-04-12
travis configuration for freemarker

pradeepmurugesan:


build.xml
```diff
@@ -133,6 +133,20 @@
     <mkdir dir="build"/>
   </target>
 
+  <property name="ivy.install.version" value="2.3.0" />
+  <condition property="ivy.home" value="${env.IVY_HOME}">
```
ddekany: If we utilize the `IVY_HOME` then I think we run into two problems. One is that `download-ivy` will try to overwrite the jar in it. The other is that the jar in `IVY_HOME` is certainly not part of Ant's classpath, so it won't be able to use the ivy tasks. So I guess we just shouldn't use `IVY_HOME`.

pradeepmurugesan: Yeah.. got the issue now.. So we just download our version of the ivy library and put it into the .ant/lib directory so that ant will pick it up.

correct me If I am wrong.

ddekany: Right, that's the plan. (Of course, running `download-ivy` is always done manually and is optional.)

PR #22, opened 2017-04-16
added javadoc task as part of the build process

pradeepmurugesan:


ddekany:
This doesn't run `test` as far as I see (same problem with the previous version). Or does it (and then why)?

That `javadoc` calls `compile` is implementation detail (which can change in future), so it should be still called explicitly. And BTW `jar` should be also called so that `bnd` is checked. Anyway, as apparently Ant redoes dependency tasks if we call it like `ant jar test javadoc`, so I have just committed a `ci` task. Just class that.

Also, why's main stuff under `install:`? Isn't that mean to be under `script:`? (I guess `download-ivy` should be under `install:`.)

pradeepmurugesan:
@ddekany changed as per the comments

PR #27, opened 2017-07-13
FREEMARKER-55: Add default tld sources

woonsan:


ddekany:
Shouldn't these be the initial values inside `TemplateFactory.Builder`? Then as a user of the `Builder` API you don't need to set the properties to them explicitly. They are the default values after all.

woonsan:
@ddekany Thanks for the comment! Indeed it's better to put those defaults in Builder. I've updated PR with a new commit: 73a53e70a5c838af451605380fc446e55b75762c

PR #31, opened 2017-08-06
FREEMARKER-55: spring.ftl marco lib support

woonsan:
- Support <#import "/spring.ftl" as spring> like Spring Framework does.
- By default, the system lib from /spring.ftl is read from the specific classpath, not from app's template path.
   The system template aliases map can be customized through SpringResourceTemplateLoader.systemTemplateAliases property.
- The same macros and functions are defined in /spring.ftl as Spring Framework's, with syntax changes to comply with FM3.
- Note: As the system template lib support is handled by SpringTemplateLoader in this PR, it means developers should always use SpringTemplateLoader directly or indirectly in order to use the system macro library. Please review this decision.
- TODOs: review/test/refine each macro and functions in spring.ftl.


woonsan:
Closing this as we want to achieve the same goal with model object(s) in Java rather than library .ftl files.

PR #43, opened 2018-01-20
Test fails with locale using comma as decimal separator

mauronunez:


ddekany:
This is hopefully fixed by https://github.com/apache/incubator-freemarker/commit/7a94e59d908942ac8e7322dbea0986c7e5ca93cc, without modifying the template.

PR #45, opened 2018-01-24
emphasize difference between eval and interpret built-ins in the documentation

fvanderwielen:
I have been struggling with the eval and interpret built-ins, and during this I learned an essential difference between the two is that eval only evaluates expressions, and interpret accepts full-fledged templates as their input strings.

Therefore I would like to add this information to the documentation, such that others realize this sooner and find even more ease and enjoyment in using freemarker :)


ddekany:
Thanks, good point!

PR #60, opened 2019-09-19
 Make `js_string` more safety.

tokuhirom:
I want to escape the `'` and `"` in hex way.

For example, there's a template like this.

    <!doctype html>
    <html>
        <#assign v='"<svg/onload=alert(3)>'>
        <body foo="${v?js_string}">
        </body>
    </html>

Rendering result is the following. This HTML shows alerting dialog when you open it on the web browser.
It means an attacker can use this for stealing a cookie.

    <!doctype html>
    <html>
        <body foo='\'<svg/onload=alert(3)>'>
        </body>
    </html>

I know it's a programming issue. Users should use `?html` filter in this context.
But some programmers sometimes forgot to apply the filter.

The freemarker can reduce the security risk with this patch.

Rendering result with this patch is here:

    <!doctype html>
    <html>
        <body foo='\x22<svg/onload=alert(3)>'>
        </body>
    </html>


ddekany:
If we are talking about security, you should use auto-escaping (https://freemarker.apache.org/docs/dgui_quickstart_template.html#dgui_quickstart_template_autoescaping). Then such a bug in a template wouldn't be a security issue as well (you just end up with a broken attribute value). I'm not convinced that we should change the default behavior of `js_escape` for this (maybe some wouldn't like the new, long result). I mean the solution is there, and it's a safer practice. (Because, like, what if the template author doesn't quote the attribute value? Then this workaround won't work.)

JasmineXieCCD:
May I know what's the new release version and target date to support jakarta namespace change for servlet-api (javax.servlet) and el-api (javax.el)
Such as File:
https://github.com/apache/freemarker/blob/2.3-gae/src/main/java/freemarker/cache/WebappTemplateLoader.java
https://github.com/apache/freemarker/blob/2.3-gae/src/main/java/freemarker/ext/jsp/FreeMarkerJspApplicationContext.java

ddekany:
Jakarta namespace change is not related to this PR. But since I'm already typing... no plan was made in that regard. So I don't know when. There are higher priority items ATM for sure. And of course it can't just be changed in those classes, instead we need to copy all the related classes, and replace the package names in it. Maybe it should be in a separate jar, actually. This madness where everything is packed into freemarker.jar comes from the times when there was no Maven or similar.

PR #62, opened 2019-12-09
add unsafe method java.security.ProtectionDomain.getClassLoader

galusben:
See
https://ackcent.com/blog/in-depth-freemarker-template-injection/

### DIFF
--- a/src/main/resources/freemarker/ext/beans/unsafeMethods.properties
+++ b/src/main/resources/freemarker/ext/beans/unsafeMethods.properties
@@ -96,3 +96,5 @@ java.lang.System.setOut(java.io.PrintStream)
 java.lang.System.setProperties(java.util.Properties)
 java.lang.System.setProperty(java.lang.String,java.lang.String)
 java.lang.System.setSecurityManager(java.lang.SecurityManager)
+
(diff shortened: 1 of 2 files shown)


ddekany:
I close this, as it's not this simple to address this problem. Anyone interested in the status of this, see this: https://issues.apache.org/jira/browse/FREEMARKER-124

galusben:
I know it is not simple to address, but it adds a quick fix to something many are exposed to. I agree that the trust level of template authors shall be as the level of source code writers, but from the file src/main/resources/freemarker/ext/beans/unsafeMethods.properties it seems that there are some blacklisted methods.

I understand that this list is not serious protection, but it will help some people that have made the mistake of trusting someone they shouldn't with a template.

Since this blog is out there, I would strongly recommend to add this method to the blacklist. There is zero cost in doing that, and the benefit can be saving someone's ass (even tough this someone did not know what they were doing). 

@ddekany

ddekany:
The recipe in that blog entry won't work anymore with the next release (2.3.30). But it's addressed differently (and some other methods will become banned as well).

galusben:
Cool. Thanks!

PR #63, opened 2019-12-12
2.3.29 maintenance

DreamSkyOk:


DreamSkyOk:
add new field pathClassicCompatible for control singly the paths parameter of include or import whether classicCompatible.

ddekany:
Looks like a quite special need. Can you tell about the use case?

DreamSkyOk:
@ddekany 
> Looks like a quite special need. Can you tell about the use case?

for this, when use classic_compatible model, the #include in ftl file also can use relative path. I think this need maybe a normal need.

the original src:
```java
freemarker.core.Environment#toFullTemplateName{
    if (isClassicCompatible() /* FM1 only had absolute names */ || baseName == null) {
            return targetName;
    }

    return _CacheAPI.toRootBasedName(configuration.getTemplateNameFormat(), baseName, targetName);
}
```

ddekany:
What I meant is that using `classic_compatible` in 2020 is already a quite special need (or I hope so), and then, using that, but not expecting exactly the classic behavior is even more so. So, why are you using `classic_compatible`?

DreamSkyOk:
@ddekany 

for using freemarker simply but having exeception "The following has evaluated to null or missing".
And I have found that Using TemplateExceptionHandler can resolve my question.

PR #66, opened 2020-04-30
Linking project overview dashboard to README

alexkarezin:
Adding a link to the README header. The linked dashboard describes overall repository structure. The goal is to help new developers explore project and understand architecture/dependencies. Direct link: https://sourcespy.com/github/apachefreemarker/


ddekany:
I would refrain from promoting development tools, especially on a place associated with the ASF, unless they already had significant role in the development.

Also, I think this doesn't help users to find their way in the case of this project much. To be fair, that's maybe in big part because of legacy that's only kept for backward compatibility, but shown prominently by this tool.

PR #67, opened 2020-05-21
Implement a #ftl onlyIncludeOnce setting.

rototor:
Using this setting allows files to be only included and interpreted once.

This is the freemarker equivalent of MS C++ ```[#pragma once]```. Which was a fix for
the common C-header pattern:

```
[#ifndef MYFILE_H]
[#define MYFILE_H]

... rest of myfile.h

[#endif]
```

This looks in freemarker like that:

```
<#if !(MYFILE_INCLUDED!false>
<#assign MYFILE_INCLUDED=true>

... rest of myfile
</#if>
```

which of course is also very ugly.

As a freemarker project gets bigger and more complex many lib templates
containing only macros and functions arise. And the same old C++ problems with it:
Those libs depend on each other, some include the required other libs explicit,
some just assume they have already be included. Which of course
causes problems from time to time.

Using ```<#ftl onlyIncludeOnce=true>``` allows to always explicit specify the
dependencies without getting a runtime penalty or some unwanted side effect
(e.g. reset of global variables) from it.

The name onlyIncludeOnce should match the other setting names. But feel free to suggest another name for the setting. 

I'm also not quite sure how to do a test for this here. I used the ExampleTest as base for this, as it just did what I needed. But it's from the manual package, and I'm using it in core here... Not really that clean. 

Note:  I already have a Apache CLA signed.


ddekany:
`#import` has that feature (among others), and is generally recommended if you need many macros/functions. Was that considered?

rototor:
Using ```#import``` in the first place would have made sense. But changing our 65k lines freemarker template codebase, which built up the last 10 years,  to use ```#import``` instead of ```#include``` is not gonna happen... As you must change all affected places to use the namespace defined by ```#import```. And ensure that the namespaces are named exactly the same in every file. Otherwise you get in trouble as soon as you search for something...

There should at least be a note in the ```#include``` documentation, that you should seriously consider using ```#import``` if you are including macros/functions. And maybe even an option in the ```#ftl``` directive to specify a "import only" file. Also some way to specify and enforce the "target" namespace name would be helpful.

This pull request is more a fix/workaround for codebases you simply can't change to ```#import```.

ddekany:
The idea of `#include` was simply to behave as if you have copy-pasted the other template there. That's a very simple concept to understand for the users. But it becomes unwieldy if you build a complex systems of templates (which I don't think was expected by the authors originally). So namespaces and `#import` was added. Even with the "only include once" feature, the reusable files will have no own namespace, so they aren't well separated, which can be a maintenance problem. 

I don't think we should support this pattern, which we don't want users to follow, in the core (which is already quite bloated). How about you create a `TemplateDirectiveModel` , and use it like `<@includeOnlyOnce>... your current template content comes here ...</@includeOnlyOnce>`? It could get the template name from the `Environment`.

Some notes, both for the current implementation, and for the possible `includeOnlyOnce` implementation:

* You should suppress output printed inside the template that's "only included once". Surely you don't want that, as it should be transparent if something is included for the first time or not.

* "Only once" should mean once per namespace, and not once per `Environment`.

* Deciding based on `Template` object identity if something was already ran won't be good enough, because `Configuration.getTemplate` might returns a different instance for the same name if the cache behind was evicted (also there's actually no hard guarantee that it caches at all). You have to use the full name of the included template instead.

rototor:
A "includeOnlyOnce" TemplateDirectiveModel would not work well, as there is now way for IntellJ IDEA (which has a rather good Freemarker support) to understand what this directive would mean. And having good IDE support is important for us.

The more I think of it the more I see that implementing this for ```#include``` is wrong. We should really use ```#import```. Even if migrating our codebase from "all stuff is in the global namespace" to namespaces per ```#import``` will be painfull.

ddekany:
It was a good point though that the documentation of `#include` should push people towards `#import`, so I have updated it accordingly: https://freemarker.apache.org/docs/ref_directive_include.html#ref.directive.include

PR #69, opened 2020-07-21
Update the OSGi BREE (minimum Bundle Required Execution Environment) to 1.7

mbooth101:
It looks like Freemarker requires Java >= 1.7 these days, the OSGi metadata should be updated accordingly.


ddekany:
Indeed, thanks!

PR #70, opened 2020-10-02
OSGI: Allow FreeMarker being imported to modular Java projects (Fix #FREEMARKER-159)

TrackerSB:


TrackerSB:
Just to mention: I tested the modification with my own modular Java project and it works. After the modification any modular Java project can require FreeMarker via `requires org.freemarker`.

PR #72, opened 2020-11-13
Add a TravisCI job to run the build and tests on ARM64 architecture

martin-g:
More and more software deployment and deployment is being done on ARM64 CPU architecture.
It would be good if Apache FreeMarker is being regularly tested on ARM64.

I've changed the way OpenJDK and Ant are installed because they are different versions on AMD64 and ARM64 VMs at TravisCI. Now they are installed manually but this way it is easier to install specific/newer versions when needed. Let me know if you prefer the old way and I will revert to it!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ addons:
 
 cache:
   directories:
-    - $HOME/.ivy/cache
+    - $HOME/.ivy-freemarker/cache
 
 before_install:
   - lscpu
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,19 +22,16 @@ arch:
   - amd64
   - arm64
 
-install:
+addons:
+  apt:
+    packages:
+      - openjdk-8-jdk
+      - ant
+
+before_install:
   - lscpu
-  - ARCH=`uname -p`
-  - echo $ARCH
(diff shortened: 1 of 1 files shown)


emiliofernandes:
At $$$dailyJob we started using AWS Graviton2 and so far we didn't have any issues with FreeMarker but it would be good to know that it is regularly build and tested on ARM64!

martin-g:
Any comments on the suggested changes ?

ddekany:
I'm not sure why's it a problem that the two Ant versions differ. I think that's actually beneficial in testing, as far as both is higher than the minimum required in the README. Installing exact version as it's in the PR now is more likely to break over time.

martin-g:
@ddekany I've re-worked it to use `addons.apt.packages` which installs the JDK and Ant from Ubuntu repositories.

martin-g:
I've also added
```

cache:
  directories:
    - $HOME/.ivy/cache
```
but I have the feeling this didn't help. Do you know I should use a different folder instead ?

ddekany:
The build uses its own Ivy cache, because `ivysettings.xml` adds resolvers. That cache is by default in the project directory, which is a problem, as CI-s typically drop and check that out for each build. So, I have just pushed a change, and now the `ci` Ant task will use `$HOME/.ivy-freemarker/cache`. So please try to cache that.

martin-g:
@ddekany This fixed it! Now it re-uses the cached jars!

ddekany:
Thanks, merged!

PR #74, opened 2021-02-23
Fixed deprecated method that caused Spring loading of Configuration to fail

j256:
Related to: https://issues.apache.org/jira/browse/FREEMARKER-177

### DIFF
--- a/src/main/java/freemarker/core/Configurable.java
+++ b/src/main/java/freemarker/core/Configurable.java
@@ -2900,19 +2900,12 @@ public class Configurable {
     }
     
     /**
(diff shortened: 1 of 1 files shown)


ddekany:
In case you don't get notification about that, note that I have commented on the Jira issue. I can't recreate the issue, and I want to, to explore alternatives.

j256:
Sorry for the delay.  I've added a comment to the Jira.  This seems is not longer an issue since they made changes in Spring version ~4.X.X.

ddekany:
I'm closing this then.

PR #77, opened 2021-11-09
Update malformed.tld

powercomstar:
the <ta> should be <tag>, it is correct


ddekany:
That's an intentionally malformed XML. However, it's not actually used in any tests, so that file should be just deleted.

PR #82, opened 2022-06-13
replace String by StringBuilder in ElementModel

GVDV:
I had a problem when using Freemarker with xml 's containing large content ( > 600KB)
The number of nodes were high, more then 10 000 and when using String concatenation as in the following snippet the performance was/is very poor.

 _if (nodeType == Node.ELEMENT_NODE) {
                String msg = "Only elements with no child elements can be processed as text."
                             + "\nThis element with name \""
                             + node.getNodeName()
                             + "\" has a child element named: " + child.getNodeName();
                throw new TemplateModelException(msg);
            } else if (nodeType == Node.TEXT_NODE || nodeType == Node.CDATA_SECTION_NODE) {
                result += child.getNodeValue();
            }_

For a xml file of 600kb this took 6 to 11 seconds to process.
When using a StringBuilder this takes 60 ms.

### DIFF
--- a/src/main/java/freemarker/ext/dom/ElementModel.java
+++ b/src/main/java/freemarker/ext/dom/ElementModel.java
@@ -131,21 +131,21 @@ class ElementModel extends NodeModel implements TemplateScalarModel {
     @Override
     public String getAsString() throws TemplateModelException {
         NodeList nl = node.getChildNodes();
-        String result = "";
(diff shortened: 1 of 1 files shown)


ddekany:
Yeah, that was a quite bad implementation there. Thanks! On the other hand, even if the file is huge, I would expect one big text node, unless there were lot of comments slicing it up.

PR #83, opened 2022-08-25
Add fine-grained mixed-content support to markup output formats.

nolaviz:
The existing logic is all-or-nothing: either an output format permits mixed content, in which case output of any other format may be interspersed in it, or it does not - and mixed content is completely disallowed. (The latter is recommended, as mixed content often indicates an application bug).

This PR enables an output format to enable mixed content while implementing an `outputForeign` method, which can inspect the foreign output's MIME type, and either pass it through, block it, or even output the foreign-format content using the new format. This e.g. enables a custom format to embed (and custom-escape) HTML that was provided in the data model as `TemplateHTMLOutputModel` rather than a plain string.

### DIFF
--- a/src/main/java/freemarker/core/CombinedMarkupOutputFormat.java
+++ b/src/main/java/freemarker/core/CombinedMarkupOutputFormat.java
@@ -70,6 +70,11 @@ public final class CombinedMarkupOutputFormat extends CommonMarkupOutputFormat<T
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


ddekany:
Do you have an ICLA (Individual Contributor License Agreement) at Apache?

nolaviz:
Yes, I have an ICLA (under my @google.com email address).

ddekany:
OK. However, I can't only check if I know the full address, so please send that to private@freemarker.a.o (replace a.o with the real domain)! Only the project managers will see it. Thanks!

ddekany:
OK, got the ICLA. (Never mind my deleted message.) Will double check if it's how we want this, then merge.

PR #88, opened 2023-01-20
Support "forced" auto-escaping policy.

nolaviz:
When the "forced auto-escaping" policy is in effect, only output formats that support auto-escaping may be used; also, the `?no_esc` built-in and the `<#noautoesc>` directive are disabled.


ddekany:
Looks good at quick glance. (Will review again before it's actually merged.) One thing comes to my mind is that the error messages should probably refer to the "auto_escaping_policy setting", or else users won't know what to do.

ddekany:
I have also done a few follow up commits to refine this.

PR #89, opened 2023-02-03
Add a simple mechanism for runtime debugging of templates.

nolaviz:
This PR adds a `TemplateProcessingTracer`, which can be used for runtime non-intrusive debugging of templates (e.g. collecting logs or coverage maps of executed lines, or profiling template execution performance).

### DIFF
--- a/src/main/java/freemarker/core/Environment.java
+++ b/src/main/java/freemarker/core/Environment.java
@@ -188,6 +188,8 @@ public final class Environment extends Configurable {
 
     private boolean fastInvalidReferenceExceptions;
 
+    private TemplateProcessingTracer currentTracer;
+
     /**
      * Retrieves the environment object associated with the current thread, or {@code null} if there's no template
      * processing going on in this thread. Data model implementations that need access to the environment can call this
@@ -301,6 +303,13 @@ public final class Environment extends Configurable {
     }
 
     /**
+     * Sets the tracer to use for this environment.
+     */
+    public void setTracer(TemplateProcessingTracer tracer) {
+        currentTracer = tracer;
+    }
+
+    /**
      * Processes the template to which this environment belongs to.
      */
     public void process() throws TemplateException, IOException {
@@ -311,12 +320,14 @@ public final class Environment extends Configurable {
             clearCachedValues();
             try {
                 doAutoImportsAndIncludes(this);
+                if (currentTracer != null) currentTracer.start();
                 visit(getTemplate().getRootTreeNode());
                 // It's here as we must not flush if there was an exception.
                 if (getAutoFlush()) {
                     out.flush();
                 }
             } finally {
+                if (currentTracer != null) currentTracer.end();
                 // It's just to allow the GC to free memory...
                 clearCachedValues();
             }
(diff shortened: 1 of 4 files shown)


chrisrueger:
@nolaviz this looks very interesting. What comes to mind is using this mechanism to detect "template coding mistakes" (even from well itentioned users). E.g. we always thought about how to hook into <#list> processing to detect "slow" [sequence concatenation](https://freemarker.apache.org/docs/dgui_template_exp.html#dgui_template_exp_sequenceop_cat) which we can use to suggest improvements to the template authors. 

Do you think your tracer would make this  somehow possible?

What is your motivation or use case?

ddekany:
I don't know... Seems to me that the functionality added is not very universal/flexible, because all you know are the positions in the source code. So it's only good for a kind of coverage heat map, and related performance analysis. Maybe that's the best we can achieve with keeping both backward compatibility and our sanity, but I will think about it a bit more, eventually.

nolaviz:
I've actually used this for much more than just coverage: e.g. I have a "watchpoint" mechanism using this infrastructure (this lets me log the values of variables when the template reaches a specific point), as well as the reverse mechanism - logging the line(s) where a certain condition evaluates to "true".

It may not seem like a lot, but when debugging production issues on a system that processes 200K requests/second and can suffer no downtime, this is invaluable.

ddekany:
Indeed, you can call `Environment.getCurrentEnvironment()`, and that opens up new possibilities. Maybe `Environment` should be passed as argument to advertise that fact more.

chrisrueger:
How are the chances to get this in the next release? @ddekany

ddekany:
This is in the TODO for the next release. Will see if on closer look I see a better approach. But my current guess is that it will be in it. (The plan is that I will work on FM near the end of the year.)

chrisrueger:
Great, thanks. One more question maybe on the two of you @nolaviz and @ddekany :
Does this mechanism allow to also hook into processing of lambda-ish expressions ([testcase](https://github.com/apache/freemarker/pull/89/files#diff-a9597da42cf027f502e4d4736ec8e2cf507690c200159ab7c1b420f5de2ddfbdR35) shows `<#list>`)? 

e.g. 
- `${seq?join(",")}`
- `${seq?filter(el -> el % 2)?map(el -> ("foo" + el))}`

ddekany:
No, it can't trace lambda-ish things, as it doesn't monitor expression evaluation at all. The tracer object only called before and after "elements", like directives, interpolations, a static text blocks. If it was called for each single expression AST node, I wouldn't feel convenient about that, because 99.9...9% of users will not use this feature ever, but the extra `if (tarcet != null) { ... }` branch is there for everyone. Yes, you can be pretty sure that it's just speculatively skipped, and should have no practical performance impact, but still... who knows, both the JIT and the modern CPU-s are complex. And more importantly, since we only provide position information (and no more detailed semantics), it has limited value to go into ever more details. Probably what you care about on expression level is Java method execution speeds, but monitoring that is an `ObjectWrapper` concern.

ddekany:
I have refined this quite heavily. See https://github.com/apache/freemarker/commit/52992c34c09a477f23d32b19703e4de8a9bfc805, and of course, tell if you see any problems with it.

PR #90, opened 2023-02-24
FREEMARKER-214 - updating JavaCC to 7.0.12

donnchadhpop:
FREEMARKER-214

Updating to JavaCC to 7.0.12 to avoid allocating new java.lang.Error instances and their associated stacktraces for each FMParser instance.

https://issues.apache.org/jira/browse/FREEMARKER-214

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -155,11 +155,6 @@
     <delete dir="build/javacc-home.tmp" />
 
     <replace
-      file="${_javaccOutputDir}/FMParser.java"
-      token="private final LookaheadSuccess"
-      value="private static final LookaheadSuccess"
-    />
-    <replace
       file="${_javaccOutputDir}/FMParserConstants.java"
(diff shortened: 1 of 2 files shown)


donnchadh:
We’ve deployed a build of our platform using a build of freemarker with these changes and we’ve seen a 50% reduction in the size of our Java Flight Recorder recordings, which we’re finding very helpful.

donnchadhpop:
We’ve deployed a build of our platform using a build of freemarker with these changes and we’ve seen a 50% reduction in the size of our Java Flight Recorder recordings, which we’re finding very helpful.

ddekany:
It's practically sure that I will review this to make it part of the next release, no worries. I just don't know when that will be, but, you won't have the re-patch FreeMarker (assuming I won't find any problems, and I don't expect any).

PR #93, opened 2023-05-10
Update docgen.cjson - fix broken link to template tester

prushforth:


ddekany:
Thanks for reporting this! Was kind of painful to realize... :)

PR #95, opened 2023-07-10
FREEMARKER-218 - Add jakarta.servlet support

dtrunk90:


ddekany:
Something like that should work. At quick glance, the equivalent of `FreemarkerServlet` is missing. So far there was only one of that, but now even the classes that used are moved into the `jakarta` package (I assume).

ddekany:
Related discussion on the dev mailing list: [https://lists.apache.org/thread/s15x9gzyhpr9d7cj8ck6kzgz3l6nfjrf](https://lists.apache.org/thread/s15x9gzyhpr9d7cj8ck6kzgz3l6nfjrf)

If this approach wins (having a new package inside the usual `freemarker.jar`), I think we will generate these new packages, rather than maintain them manually. Will see... Probably, that's best to address after the Gradle PR was merged.

dtrunk90:
> Related discussion on the dev mailing list: https://lists.apache.org/thread/s15x9gzyhpr9d7cj8ck6kzgz3l6nfjrf
> 
> If this approach wins (having a new package inside the usual `freemarker.jar`), I think we will generate these new packages, rather than maintain them manually. Will see... Probably, that's best to address after the Gradle PR was merged.

FYI: For gradle there's https://github.com/nebula-plugins/gradle-jakartaee-migration-plugin which could probably help generating the new package.

ddekany:
What features of Servlet/JSP support are you using? I'm asking because support JSP taglibs (TLD-based) might brings some complications, if that has to work on modern containers.

PR #97, opened 2023-10-29
Speedup ConcatenatedSequence iteration

chrisrueger:
We managed to get improved performance for scenarios where after creating a list and then iterating over it. 

* store the size at creation time so that we avoid repeated expensive calculation in get() or size().
* this speeds up scenarios where ConcatenatedSequence are iterated like BuiltInsForSequence like ?join or <#list>

```
<#assign mylist = [] >
<#list 1..10000 as i><#assign mylist = mylist + [i+"foo"!?upper_case] /></#list>

${mylist?join(",")}

or

<#list mylist as e>
${e}
</#list>

```

Note:
After the change this code finishes in `< 1s`.  Before it almost never finished...

### DIFF
--- a/src/main/java/freemarker/core/AddConcatExpression.java
+++ b/src/main/java/freemarker/core/AddConcatExpression.java
@@ -196,22 +196,29 @@ final class AddConcatExpression extends Expression {
         TemplateSequenceModel {
         private final TemplateSequenceModel left;
         private final TemplateSequenceModel right;
+        private final int leftSize;
+        private final int totalSize;
 
-        ConcatenatedSequence(TemplateSequenceModel left, TemplateSequenceModel right) {
+        ConcatenatedSequence(TemplateSequenceModel left, TemplateSequenceModel right) throws TemplateModelException {
             this.left = left;
             this.right = right;
+            // Calculate sizes only once and reuse them below
+            this.leftSize = left.size();
+            this.totalSize = leftSize + right.size();
+            
         }
 
         @Override
         public int size()
         throws TemplateModelException {
-            return left.size() + right.size();
+        	// No more expensive computations here
+            return totalSize;
         }
 
         @Override
         public TemplateModel get(int i)
         throws TemplateModelException {
-            int ls = left.size();
+            int ls = leftSize;
             return i < ls ? left.get(i) : right.get(i - ls);
         }
     }
--- a/src/main/java/freemarker/core/AddConcatExpression.java
+++ b/src/main/java/freemarker/core/AddConcatExpression.java
@@ -22,6 +22,7 @@ package freemarker.core;
 import java.util.HashSet;
 import java.util.Set;
 
+import freemarker.template.Configuration;
 import freemarker.template.SimpleNumber;
 import freemarker.template.SimpleScalar;
 import freemarker.template.SimpleSequence;
@@ -36,6 +37,7 @@ import freemarker.template.TemplateNumberModel;
 import freemarker.template.TemplateScalarModel;
 import freemarker.template.TemplateSequenceModel;
 import freemarker.template._ObjectWrappers;
+import freemarker.template._VersionInts;
 
 /**
  * An operator for the + operator. Note that this is treated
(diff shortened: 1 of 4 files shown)


ddekany:
Earlier we weren't assuming that the sizes won't change between the concatenation and the iteration; the concatenated sequence is just a "view". I know that happens almost never in practice, but still, this should be at least `incompatibleImprovements`-dependent behavior, as it's not backward compatible, and the old behavior is not bug (but a bad design decision). Also, if from now on we do assume that the size of underlying sequences won't change, then certainly there's a more efficient solution (like where reading the 1st element is not O(N), but O(log2(N))).

ddekany:
Yeah, and we got the same historical issue with "hashes" (i.e. `Map`-s). I'm not saying that they have to be fixed at once, just that these union views are one of the long standing performance traps that people keep walking into.

chrisrueger:
> Earlier we weren't assuming that the sizes won't change between the concatenation

I feared that was the case 😄 

If we are speaking incompatible improvements: What else would be possible? 
Do you think we could even discuss e.g. making the concatenation configurable e.g. via a callback via Configuration.java, so that we could plugin our own concatenator (sequence or hash)? 
That way more control would be possible but also backwards-compatibility, since we could keep the existing behavior.

ddekany:
I hope that for something technical like this, we can provide a default behavior that's good enough for everybody. (Except for those who managed to exploit the "view" nature of the old behavior... which is unlikely, but we promise that 2.3.x-s can be updated without checking the changelog.) Then the implementation can be kept internal, and we switch it if `incompatibleImprovementes` >= 2.3.33.

chrisrueger:
> `incompatibleImprovementes` >= 2.3.33.

Ok I will see if I can add that.

chrisrueger:
I added a check for `incompatibleImprovementes >= 2.3.33`

How would you go about optimizing reading the 1st element?  @ddekany

ddekany:
How to accelerate this further: My idea is this. We can accelerate accessing elements sequentially, by index, with keeping the current data structure. The current data structure is (assuming I remember correctly) such that the concatenation result is a TemplateSequenceModel, which is a the root node of an unbalanced binary tree, where each nodes itself is TempalteSequenceModel that stores a left- and right-sequence (and with this PR now also their sizes). We could maintain a cursor in the root node. This cursor stores the path in the tree, from its root, to the last element that we had to retrieved. The path is just a List of elements, where each element stores a reference to a tree node, and the left-most index of the sequence slice the node represents in the whole (root node) sequence.  When a new element is get by index, we can navigate to it starting out from that cursor, and not from the root node. The cursor initally points to the root node. Then, accessing seq[0] is O(N) for the first time, but O(1) next time, as the cursor (the last element of the path) already points that element. Furthermore, accessing the next element (seq[1]) will also be O(1), since if the next index is out of the range for the current node, you just have to step back in the path (hence naviating to the parent node), and then down to the right node.

chrisrueger:
@ddekany thanks for the algorithm. I am just trying to wrap my head around this and writing a test. Could you please describe an example template which would benefit from this optimization? Just want to make sure I am thinking in the same direction as you.

ddekany:
The cursor thing meant to be beneficial if someone (or something) gets the sequence elements by index, because, usually, you access adjacent (or even just nearby) indexes. But I was just thinking about the data structure that we have, and how to avoid it being surprisingly slow for the said seemingly basic use case. Maybe it's not something that happen in templates? I did not dig into this honestly...

chrisrueger:
@ddekany ok I see. I would suggest to do more optimization in future PRs. This current optimization which I have put in place solves a big issue when it comes to iteration, which would be great to have it in.

ddekany:
I will dig into this more, probably in the coming days, and maybe I will improve it. I guess this will be in the next release in some form.

ddekany:
I have realized that this can be done in a backward compatible enough way, if I support creating a `TemplateModelIterator` (which `#list`, `?join`, etc will prefer if available), then implement an `TemplateModelIterator` that just traverses the tree deeph first, left to right. As most users never set `incompatibleImprovements`, this is a significant advantage, as this way they will benefit from it as well.

See https://github.com/apache/freemarker/commit/93a96377318e5775418de24ff60dfae25d54f93e

If you agree, I would close your PR. While my solution doesn't speed up indexed access, I want to work on separately from this, as the current commit already fulfills the stated issue (slow iteration).

chrisrueger:
@ddekany great to hear, thanks.

> If you agree, I would close your PR.

Yeah sure. Just close it.

chrisrueger:
Closing as alternative solution exists now.

ddekany:
Tweaked it further, for `size()` and `get(int)`, still in a way that doesn't need `incompatibleImprovements` (so it's not the fastest, but at least everyone gets it automatically):

https://github.com/apache/freemarker/commit/d75ddc6c9bebd4fdb1b332954b619cfda704c5a1
https://github.com/apache/freemarker/commit/8ac81c610a12d461eca87ae92ea91284ca089ae4

PR #98, opened 2023-11-04
[FREEMARKER-224] change Automatic-Module-Name to org.apache.freemarker

bmarwell:
fixes [FREEMARKER-224](https://issues.apache.org/jira/browse/FREEMARKER-224)

As the current module name is not released yet, this is not a breaking change.
For further information and motivation, please see the issue.

This, of course, is open to discussion.
However, I strongly recommend to follow the reverse domain name system. Additionally, the is no need to have the root package and the module name match.

### DIFF
--- a/osgi.bnd
+++ b/osgi.bnd
@@ -63,4 +63,4 @@ Specification-Vendor: freemarker.org
 Implementation-Title: FreeMarker
 Implementation-Version: ${versionForMf}
 Implementation-Vendor: freemarker.org
-Automatic-Module-Name: freemarker
+Automatic-Module-Name: org.apache.freemarker


bmarwell:
Discussion: the website uses `freemarker.apache.org` which also makes it a good choice.

ddekany:
But it was released with "Automatic-Module-Name: freemarker" already. The module name was "freemarker" because, as far as I remember, in Java 9, if you had no explicit module name, it was deduced from the jar name, and was "freemarker". (Also that happens to be the top-level Java package name.)

bmarwell:
> But it was released with "Automatic-Module-Name: freemarker" already.

Oh, dang. I missed that. :( 

> The module name was "freemarker" because, as far as I remember, in Java 9, if you had no explicit module name, it was deduced from the jar name, and was "freemarker". 

While this is true (and still is for any later java version), this would not have been a problem and a reason to use `freemarker` as the automatic-module-name. Whenever you build something with gradle or maven which uses "requires module-name-from-file-name", it gives a BIG warning not to push something like this into the wild. Thus, no-one could have used it. The moment someone renames the `freemarker.jar` locally, it would break.  Hence, having set the module-name to something else would not had hurt back then.

> (Also that happens to be the top-level Java package name.)

See issue. for JPMS, it is even more strongly encouraged to use a reverse domain name.

So then I will rebase this to the 3-branch. Or maybe even to the 2.4-branch, if you agree? WDYT?

ddekany:
Whatever warnings tools give, people will still import "freemarker", and they rightfully unhappy, if for some reason FreeMarker micro version increases, and things suddenly break. Anyway... it's just not worth the energy.

As of "3", I believe there we simply need {{module-info.java-s}} in all projects. (Not sure about OSGi support, and if we will support it at all, or it will be practically dead by the time FM3 released, if ever.)

Actually, as of more practical problems, I suspect 2 doesn't even work properly when used as a module, as it should static import a few stuff for it, like logging frameworks (and even the module of JUL). So if that suspicion is right, and it's actually broken in that regard, then even in 2 will need a {{module-info.java}}.

bmarwell:
> Actually, as of more practical problems, I suspect 2 doesn't even work properly when used as a module, as it should static import a few stuff for it, like logging frameworks

That's not a concern. Automatic modules will get access to everything and export everything. You can't create a jlink distribution with automatic modules.

Therefore it doesn't break anything because no one could have used freemarker as a module before the version where you introduced the automatic module name.

> As of "3", I believe there we simply need {{module-info.java-s}} in all projects.

That's definitely the best solution.

> (Not sure about OSGi support, and if we will support it at all, or it will be practically dead by the time FM3 released, if ever.)

OSGi is actually more widely used compared to JPMS modules. It's better defined, more Flexible and has been around for longer. I myself don't use it. But that would be a great question on the user mailing list.

chrisrueger:
> Not sure about OSGi support, and if we will support it at all, or it will be practically dead by the time FM3 released, if ever

Oh please keep OSGi support. We use it actively since many years. I am a contributor at [bnd / bndtools](https://github.com/bndtools/bnd) I could try to help in case of OSGi related questions or testing. bnd is currently used to create the OSGI metadata in the ant build.xml and there is also similar support for maven or gradle.

bmarwell:
Hi @ddekany 

I asked in #maven on ASF slack. The general consensus is:

1. derived module name derived from file name is unstable. Gradle users might not  get the same result, so this is not a reason to name the module "freemarker"
2. package is not linked.
3. The software name is "Apache Freemarker", this should be reflected in modules and packages if possible.
It is not for the package, so do it for the module.
4. The module name should be reverse DNS

And most important:

5. The adoption rate is probably exactly 0%. There is not reason not to do a breaking change. If you like, we can ask on the users list. But JPMS projects are rare.
6. Even then, it is super easy to switch and very straight-forward.

Given some Apache Maven Members agree on `org.apache.freemarker`, I highly recommend the switch.

ddekany:
So auto modules export everything, and we should be fine. That's good to know.

I'm not seeing the relevance of the fact the module name derived from jar name is unstable. That's why it was cemented into the META-INF in some already released version(s). If someone used FreeMarker as an auto-module before that, they probably just referred to it as "freemarker", because that was its derived name. Most users couldn't care less about the module name, as far as it always remains the same, and, that's what was done.

Yes, I know how it should be, and certainly was well aware when the decision was made back then. Because as I said, it was deliberate.

I don't know the adoption rate. Somebody asked for the inclusion of "Automatic-Module-Name", so it's at least 1 person. Why contemplating it? It doesn't worth the energy, as far as I'm concerned.

ddekany:
> Oh please keep OSGi support.

That comment only applied to FM3 anyway. But, sure. BTW, in the Gradle build (of 2.x), it uses a much more fresh bnd internally, and the contributor said that it puts less packages into the "uses" sections. Anyway, when the time comes to review that, I will ask if I can't figure it out if it breaks something...

bmarwell:
> It doesn't worth the energy, as far as I'm concerned

As far as I am concerned, it is worth the "energy". 

I think this should be a vote on the mailing list then, according to Apache voting rules. We can CC the users list. Any user with an adoption not willing to change it for a 2.4.x-version is reason enough not to do it.

ddekany:
Given the trivial nature of the commit itself, is it OK if I close this? Because, I would like to reduce the number of pending PR-s to 0. Of course you can bring it up on the dev list if you think, and we also have a Jira issue for this.

PR #99, opened 2023-11-04
add .asf.yaml

bmarwell:
Adding the .asf.yaml file allows control over a few things.

* Setting the project name
* Setting the project homepage
* Auto-linking issues (like FREEMARKER-224 which would automatically turn into a link after this is merged)
* Auto-linking pull requests into the correspinding JIRA issues
* Dependabot alerts (for developers with the respective karma only)
* enabling/disabling of specific merge buttons.

I am open to any changes, I would just like to add a little bit of convenience using the jira options and auto links.

We could also add labels to the github repository if wanted.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -1,0 +1,33 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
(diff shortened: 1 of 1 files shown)


bmarwell:
@ddekany I forgot to set issues to false. You should do that before people start opening issues here.

bmarwell:
Fix:
https://github.com/apache/freemarker/pull/101

PR #100, opened 2023-11-04
[BUILD] add github workflow

bmarwell:
Just a suggestion. This way further pull requests will automatically be checked.


bmarwell:
Because of the prompt, I was not able to add the "dist" goal. However, the signing could be done by a passphrase-less gpg key. This is needed for a rat-check. Skipped it for now.

bmarwell:
Tests with other JDKs (11, 17, 21) can be added in another commit, if you are interested.

ddekany:
Thanks! BTW, Travis ran very long ago, when it was still free, but it's green build icon is still there. If there something like that Git actions, it should be replaced to with that.

bmarwell:
Sure! Can do that in another PR, preferably. Otherwise I'd do it blindly.

I would love to add the "ant dist" target. The only thing I can't do is "answer yes" on the last question.
I don't know much ant. But maybe we could add a build property which skips the interactive question on CI builds?

.github/workflows/ant.yml
```diff
@@ -0,0 +1,64 @@
...
+          echo "mvnCommand=$(which mvn)" >> build.properties;
+          echo "gpgCommand=$(which gpg)" >> build.properties;
+
+      - name: Prepare ant with ivy
+        shell: bash
+        run: ant download-ivy
+
+      - name: Build with Ant and ivy
+        shell: bash
+        run: ant jar test
```
bmarwell: I think this would be the ci goal instead

bmarwell (commented):
This PR should also remove Travis.

bmarwell:
We can add a cache in a later PR, but for now this should be fine. Maybe not even worth it as the project is heading towards Gradle.

ddekany:
We don't need `ant dist`, but `ant ci`, which I see you have discovered. If we want to test more, it should be added to `ci`. But, yes, these will go away with the Gradle PR anyway. (There's also Github actions prepared for that by the author of the Gradle PR).

PR #104, opened 2023-12-30
FREEMARKER-218: Jakarta support

kelemen:
Added new generated sources created from `freemarker-javax-servlet` where javax references are replaced with jakarta references. Tests are also copied (with similar adjustments).

The main remaining question is `freemarker.cache.WebappTemplateLoader`. For now, I have left it in the same package renamed to `JakartaWebappTemplateLoader`. There are two reasons for this:

- `freemarker.cache` is a split package, and it is awkward to deal with this, if we move this class to a new package.
- This class refers to package private classes and methods from `freemarker.cache` (`URLTemplateSource` and `TemplateLoaderUtils`), and making them public is an obvious detriment.

That said, I do have a variant of this change where I moved this class into another package (in my local copy), so moving to a new package is definitely doable, if the used classes (and their methods) are made public.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,6 +48,11 @@ jobs:
         with:
           java-version: 16
           distribution: zulu
+      - name: Set up JDK 17
+        uses: actions/setup-java@v3
+        with:
+          java-version: 17
+          distribution: oracle
       - name: Validate Gradle wrapper
         uses: gradle/wrapper-validation-action@v1.1.0
       - name: Run Build
@@ -60,4 +65,3 @@ jobs:
           name: test-reports-${{ matrix.os }}
           path: build/reports/**
           retention-days: 30
-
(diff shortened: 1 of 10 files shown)


ddekany:
My preference is to copy `freemarker.cache.WebappTemplateLoader` into `freemarker.ext.jakarta.servlet.WebappTemplateLoader`, and make `URLTemplateSource` (and most of its methods) and `TemplateLoaderUtils` public. (These are very old classes, and it's not likely that we will want to change then in an incompatible way, so we might as well let others use them for their own `TemplateLoader`-s.) My reason is exactly to move away from the split-package style. One day the whole javax.serlvet support will be removed, and only jakarta.servlet will remain, and at that point we will delete `freemarker.cache.WebappTemplateLoader`, and then the whole servlet support can be separated without spit-package issues.

kelemen:
Ok, I have pushed the variant moving it to another package.

I have made the least amount of changes to the two classes necessary. I'm assuming that if you want to make some adjustments to them you will do that after merging this PR (e.g.: making `URLTemplateSource` final).

PR #108, opened 2024-04-23
use StandardCharsets instead of string

hduelme:
Since Java 7 `StandardCharsets` can be used for standard charsets. It prevents the unnecessary lookup of the charset and in addition no  `UnsupportedEncodingException` can be thrown.


ddekany:
As this is non-trivial change, do you have a [Contributor License Agreement](https://www.apache.org/licenses/contributor-agreements.html) at Apache?

hduelme:
Good to know. I thought this is a trivial change. I will sign the CLA and notify you as soon is its done.

ddekany:
Thanks, ICLA as received.

PR #110, opened 2024-06-03
Creating multi release jar

kelemen:
The tests tasks were not touched at all. So, they do not depend on the jar. If such tasks are needed, then I think new test tasks should be added as opposed adjusting the current test tasks. This is to avoid confusing IDEs (and a considerable complications), also it would make more sense conceptually that then there would be a combined test task depending on the jar for each java version (and the tasks wouldn't be directly tied to source sets).

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ That's because different parts of the source code target different Java versions
 and Gradle requires the exact JDK version (not higher) for each.
 
 Be sure that your default Java version (which Gradle should use automatically) is at
-least 16!
+least 17!
 
 If you are building from the official source *release* (not from source that you
 got from Git), `gradle/wrapper/gradle-wrapper.jar` is missing from that, and you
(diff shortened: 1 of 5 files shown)


kelemen:
Almost forgot an important note: I had to update the bnd plugin, and this forced a dependency on JDK 17 for the build. So, now people opening the project will require to set the Gradle JVM (or `JAVA_HOME` when running via `gradlew`) to 17.

kelemen:
Actually, I have found a not too complicated way to add test tasks depending on the jar instead. Also, they are not added per source set where tests are enabled, because I have realized that they might have conflicting dependencies.

A small issue with this solution is that now the tests are run twice. While inefficient, it has a minor benefit: The failures due to the effect of the multi release jar can be separated from the test failures. Also, locally it is usually better to not run against the jar, because it is faster.

PR #111, opened 2024-07-26
Add blank_to_null, empty_to_null, and trim_to_null built-ins

chrisrueger:
- New built-in for Strings `?is_blank`
- `?is_blank` returns `true` if string is `null`, `empty` or contains only whitespaces
- See testcases
- it is similar to `somestring?trim?length == 0` (or the like), but shorter and avoids a new String object as `?trim` would produce

## Why?

It is a typical check we do and it would be nice to have this in FM - if there are no objections against it. 
I was thinking about adding `?is_empty` too (as an shorter alternative to `somestring?length == 0`) but wanted to confirm first. Let me know if that should be added too.


ddekany:
I think the better approach is to introduce a built-in or operator that normalizes blank strings to `null`. After that, all the long existing missing value operators are applicable (see: https://freemarker.apache.org/docs/dgui_template_exp.html#dgui_template_exp_missing) . Also, that targets the root of the problem, which is that you really just want either a `null`, or a non-blank `String`, but because of some technical mess in a layer out of your control, both variants can come.

As of `s?is_empty`, check `s?has_content`. Maybe there's an overlap. (Also according the above, the question can be if you want to normalize a `"  "` to `null`, or only a 0-length string.)

As of `s?trim?length == 0`, note that if `s` is `null`, that's actually an error, so that wouldn't even work. That's just one more reason for a dedicated solution of course.

chrisrueger:
> normalizes blank strings to `null`. After that, all the long existing missing value operators are applicable

Are you referring to this, that 

`${nonExisting?is_blank}` 

errors in 

`The following has evaluated to null or missing: ==> nonExisting [in template "name" at line 1, column 3] ---- FTL stack trace ("~" means nesting-related): - Failed at: ${nonExisting?is_blank}`

That is indeed something I haven't considered in my PR. 

`${nonExisting!?is_blank}`  works because of the `!`

But could you elaborate a bit more on what you have in mind? An example maybe? 

> As of `s?is_empty`, check `s?has_content`. Maybe there's an overlap.

Ah you are right. I have forgotten about `?has_content`.

As I am writing this:
I noticed that `?has_content` handles `null` already and does not fail.
e.g. `${nonExisting?hasContent?c}` prints `false`.

Should I maybe implement `?is_blank` similar to `?has_content` ?
Or are you generally against `?is_blank`?

ddekany:
What I'm saying is that the standard way of checking if `s` is `null` is `s??`, and the standard way of giving a default if `s` is `null` like `s!'my default'`. However alien they look, too late, you have to get used to them if you use FreeMarker.

And also, sometimes you want to treat a blank `s` as if it was `null`, because really it should be `null`, but the data source you receive it from doesn't normalize it. So if there was a `?blank_to_null` that normalizes blank to `null`, then this unwanted difference can be removed. And after that, `s?blank_to_null??` and `s?blank_to_null!'my default'` will just work. Only it's a bit too verbose. Back then I considered if there should be an operator to do this (like `s~`, where `~` is the operator), but that's yet another alien symbol, plus, there's the variant where you only want to normalize 0-length strings to `null`.

chrisrueger:
Thanks for your remarks. Just to add , since I have not written it explicitly: Basically I was only looking for the equivalent of Apache Commons [StringUtils.isBlank(...)](https://github.com/apache/commons-lang/blob/master/src/main/java/org/apache/commons/lang3/StringUtils.java#L3623)  but as a FM built-in. Just because it is a check people often need. 

We already know `??` and `!` and use it. Usually our templates are full of `!` (maybe not always the most elegant, but easier for most our users to ignore null and just have it an empty string) 

`someString!?is_blank` appeared to be a bit more descriptive (given the meaning of "blank" vs. "empty" is known to the template author). So somebody came to me asking if there is a `?is_blank` built-in. 
I said, no, but we could create our own `${isBlank(string)}` TemplateMethodeModel function - but a built-in seemed nicer - subjective for sure.  Just how it came to this PR .

 



> And also, sometimes you want to treat a blank `s` as if it was `null`, because really it should be `null`, but the data source you receive it from doesn't normalize it.

True, but that was not my intention with "is_blank". 

This sounds like Apache [StringUtils.trimToNull(...)](https://github.com/apache/commons-lang/blob/master/src/main/java/org/apache/commons/lang3/StringUtils.java#L9124) 

We sometimes get data (from who knows where) and we need to ask the question **in the template** "does this field contain only spaces"? 

we usually do:

`<#if firstName!?trim != "">do something</#if>`

or

`<#if firstName?? && firstName?trim != "">
do some larger stuff like creating an XML-fragment...
</#if>`

We thought that 

`<#if firstName!?is_blank>do something</#if>` is more descriptive. 

Anyway I can try adding also `?blank_to_null` if you think this should also be there. 

Sorry maybe I just thought too simplistic :)

ddekany:
My reservation is that while `s!?is_blank` is a neat trick, it only solves 1 uses cases out of the 2:

- Use-case 1: `<#if s is null or blank>`, but also, often we need the opposite, `<#if s is not null, nor blank>`
  - With `?is_blank`: `<#if s!?is_blank>`, and the opposite is `<#if !s!?is_blank>`
  - With `?blank_to_null`: `<#if !s?blank_to_null??>`, and the opposite is `<#if s?blank_to_null??>`
  - With `~` operator: `<#if !s~??>`, and the opposite is `<#if s~??>`
- Use-case 2: `get the value of s, but if it's null or blank, then evaluate to default value 'D'`
  - With `?is_blank`: It's doesn't solve the use case (or is way too verbose).
  - With `?blank_to_null`: `s?blank_to_null!'D'`
  - With `~` operator: `s~!'D'`

Admittedly in use-case 1 `?is_blank` tells more directly what it wants, but OTOH `?blank_to_null` reduces the problem to the missing value handling problem, and so will work with other existing missing value handling mechanism as well, like macro argument defaults, passing `null` to Java methods, or anything else that will be added in the future.

chrisrueger:
Ok I think I understand now. So you would prefer `?blank_to_null` over `?is_blank`, correct?

The only thing which comes to my mind is how it would play with the existing [?trim](https://freemarker.apache.org/docs/ref_builtins_string.html#ref_builtin_trim) ?

Should there also be a `?trim_to_null` and `?blank_to_null` ?

- `?blank_to_null` returns null if the string is null, empty or just whitespace. Otherwise it returns the string as is. (un-trimmed)
- `?trim_to_null` returns null if the string is null, empty or just whitespace. Otherwise it returns the string trimmed.

What do you think?

chrisrueger:
@ddekany implemented `?blank_to_null` as discussed above.  Not sure if the use of `EvalUtil.coerceModelToStringOrUnsupportedMarkup` is appropriate here or not. Let me know.

ddekany:
Yes, and then there's also `?empty_to_null`. At least I think we should have these 3. I guess then `?is_blank` is not really needed (though is still meaningful, if it does NOT tolerate `null`).

`coerceModelToStringOrUnsupportedMarkup`: It's called by `freemarker.core.BuiltInForString#getTargetString`, so to be consistent with all the other `BuiltInForString`-s, we should keep that as is.

chrisrueger:
@ddekany thanks. I added [?trim_to_null and ?empty_to_null](https://github.com/apache/freemarker/pull/111/commits/27f33c83a3bc1172fb5ddfbce9eb32b2f895831f) so we now have all 3:
- `?blank_to_null`
- `?trim_to_null`
- `?empty_to_null`

All work with `null` and `non-null`.

chrisrueger:
@ddekany added sections to the manual. I think this is ready for review from my side. Let me know, what you would like to change.

ddekany:
See post-merge follow up commit: https://github.com/apache/freemarker/commit/943c9b6030dd1e8e70db6a12a9c1a95b46e31e86

PR #114, opened 2024-08-28
Add on directive to switch as an alternative to case

scrhartley:
This is a copy of #106 which was backed out after a problem with the contributors agreement.

---

The current `switch` directive is not recommended due to its fall-through behavior being regarded as error-prone. To solve this problem, we introduce a new `on` directive as an alternative to `case`, which doesn't support fall-through. This new directive allows specifying multiple comma-separated conditions in order to address the primary motivator for using fall-through with `case`.

## Example

Using `case`:
```freemarker
<#switch animal.size>
  <#case "tiny">
  <#case "small">
     This will be processed if it is small
     <#break>
  <#case "medium">
     This will be processed if it is medium
     <#break>
  <#default>
     This will be processed if it is neither
</#switch>
```
Using `on`:
```freemarker
<#switch animal.size>
  <#on "tiny", "small">
     This will be processed if it is small
  <#on "medium">
     This will be processed if it is medium
  <#default>
     This will be processed if it is neither
</#switch>
```

## Details

* Mixing `case` and `on` directives is disallowed and will fail.
* If a `switch` contains an `on`, then using `break` or `continue` is not supported (applying to both `on` and `default`). In this situation, `break` and `continue` will follow the behavior of the containing scope, e.g. of a containing `list`. This does mean that if someone accidentally uses `break` with `on`, they could potentially become confused.

### Additional Notes

* When `on` is not used, the legacy behavior for `case`/`default` is that `continue` will be treated as `break`. If we wish to change this, then that work can be done separately.
* If using `default`, the parser also allows it to appear before or between `case` directives, but when using `on` then `default` is only allowed after.

### DIFF
--- a/freemarker-core/src/main/java/freemarker/core/On.java
+++ b/freemarker-core/src/main/java/freemarker/core/On.java
@@ -1,0 +1,95 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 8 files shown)


ddekany:
Thanks, was merged. Also I have applied exactly the same changes as after PR #106, just have put all 3 commits into 1 here.

PR #115, opened 2024-09-03
Add generics to _ArrayIterator and _ArrayEnumeration

hduelme:
I added the missing generics to  `_ArrayIterator` and `_ArrayEnumeration` and to the places they are used.


ddekany:
The reason FreeMarker doesn't use generics is that it's not backward compatible to add them to public API-s. So it couldn't adopt them. Even though TemplateElement is in principle internal, it's used by some.

hduelme:
Good to know, I thought TemplateElement was only an internal API. I removed the change to the API.

PR #119, opened 2024-11-22
Fix sep directive in syntax A breaks templates in syntax B

binokkio:
I noticed compiling a template containing `<#sep>` using a configuration specifying the square bracket tag syntax fails with a `freemarker.core.ParseException: Syntax error in template "X" in line Y, column Z: #sep must be inside a #list (or #foreach) block.` I think `<#sep>` should be ignored instead as it is not using the square bracket syntax.

This PR adds a unit test covering that case, a change in `FTL.jj` that solves the issue, and fixes a typo.

`./gradlew check` runs successfully for me with these changes, but I do not know the Freemarker codebase well enough to be confident that it does not introduce other issues.

I'd be happy to improve this PR if there is any feedback.

### DIFF
--- a/freemarker-core/src/main/javacc/freemarker/core/FTL.jj
+++ b/freemarker-core/src/main/javacc/freemarker/core/FTL.jj
@@ -685,7 +685,7 @@ TOKEN_MGR_DECLS:
         }
         
         // For square bracket tags there's no non-strict token, so we are sure that it's an FTL tag.
-        // But if it's an angle bracket tag, we have to check if it's just static text or and FTL tag, because the
+        // But if it's an angle bracket tag, we have to check if it's just static text or an FTL tag, because the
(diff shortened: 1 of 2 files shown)


ddekany:
Do you have an ICLA signed at Apache? Otherwise I can't merge code changes that are not entirely trivial. If not, I can just redo this without looking, now that the issue is known (Thanks!). Or you may prefer to sign the ICLA.

binokkio:
I don't have an ICLA, but I'm happy to sign it. Any specific instructions or shall I just follow https://www.apache.org/licenses/contributor-agreements.html#submitting ? Thanks!

ddekany:
Just follow that, yes. It's ASF-wide.

binokkio:
I have received confirmation that the ICLA has been filed in the ASF records, I think you might need the id `cwr` to find it.

PR #120, opened 2024-12-08
Re-enable bare `<sep>` when incompatibleImprovements < 2.3.34.

nolaviz:


ddekany:
(Should have been 2.3.34, not 2.3.24, but will fix that post-merge.)

PR #122, opened 2025-04-02
Disable execution for freemarker.template.utility.Execute

setuper:
This class can be used as a gadget for attack. It is suggested to make a setting to disable potentially dangerous code


ddekany:
Creating an instance of this `TemplateModel` from templates is disabled by default already (see `TemplateClassResolver.SAFER_RESOLVER`). So I'm note sure how much this helps in practice, as this doesn't block by default, and people had to realize that the problem exist at all, and then that there's a system property to block it.

PR #123, opened 2025-11-20
Fixed error in docs for new switch-on syntax

stevemcleod:
The example code for switch-on had a stray usage of switch-case. I've corrected this to consistently use the switch-on syntax.


ddekany:
Thanks! Updated website with this too.

PR #127, opened 2026-04-19
[FREEMARKER-234] Expose Boolean isFoo() getters as bean properties

mapeders:
## Summary

Fixes [FREEMARKER-234](https://issues.apache.org/jira/browse/FREEMARKER-234): `BeansWrapper` did not expose `isFoo()` methods that return `java.lang.Boolean` (wrapper) as the `foo` bean property, even though the manual states that `model.foo` should invoke either `obj.getFoo()` or `obj.isFoo()`. The restriction to primitive `boolean` came from `java.beans.Introspector`, which is strict to the JavaBeans spec.

- Gated behind `incompatible_improvements` ≥ **2.3.35** — pre-existing configurations see no change.
- `ClassIntrospector.getPropertyDescriptors` now supplements `java.beans.Introspector`'s output by scanning for `public Boolean isXxx()` methods (mirroring the existing Java 8 default-method supplement path).
- `_MethodUtil.getBeanPropertyNameFromReaderMethodName` accepts `Boolean.class` alongside `boolean.class`.
- Static `isFoo()` and non-Boolean `isFoo()` (e.g. `String isLabel()`) remain excluded from property exposure.
- Release notes entry + JavaDoc update under the "2.3.35 (or higher)" ICI change list.

## Test plan

- [x] `GetPropertyNameFromReaderMethodNameTest` — flipped the assertion that codified the bug (`isFoo` + `Boolean.class` → `"foo"`), added `isF`/`Boolean.class` symmetry.
- [x] New `BeansWrapperMiscTest.booleanWrapperIsGetterAsPropertyTest` — verifies:
  - At ICI 2.3.34 (legacy): `.obsolete` is missing and only the method form is reachable.
  - At ICI 2.3.35: `.obsolete` resolves to the wrapped `Boolean`, primitive `boolean isActive()` still works, `String isLabel()` is not exposed as a property, `static Boolean isArchived()` is not exposed as a property, and `isObsolete` remains reachable as a method.
- [x] Full core test suite: 871 tests, 0 failures (`./gradlew :test --rerun-tasks`).

### DIFF
--- a/freemarker-core/src/main/java/freemarker/ext/beans/BeansWrapper.java
+++ b/freemarker-core/src/main/java/freemarker/ext/beans/BeansWrapper.java
@@ -274,6 +274,16 @@ public class BeansWrapper implements RichObjectWrapper, WriteProtectable {
      *       templates the value can be accessed both as {@code obj.name} (like a property), and as {@code obj.name()}
      *       (for better backward compatibility only - it's bad style).
      *     </li>
+     *     <li>
+     *       <p>2.3.35 (or higher):
+     *       {@code isFoo()} methods that return {@link Boolean} (the wrapper class) are now exposed as the {@code foo}
+     *       bean property, just like {@code isFoo()} methods returning primitive {@code boolean} already were. Before
+     *       this, {@link java.beans.Introspector} (which is spec-strict) reported only primitive-{@code boolean}
+     *       {@code isFoo()} methods as property readers, so {@code Boolean isFoo()} was accessible only as a method
+     *       ({@code obj.isFoo()}), even though the documented behavior claims {@code obj.foo} should work for either
+     *       {@code getFoo()} or {@code isFoo()}. Static {@code isFoo()} methods and {@code isFoo()} methods with any
+     *       other return type are still not exposed as properties.
+     *     </li>
      *   </ul>
      *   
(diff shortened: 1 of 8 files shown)


chrisrueger:
As stated in https://issues.apache.org/jira/browse/FREEMARKER-234 this is based on the [Java Beans Specification Section 8.3.2](https://download.oracle.com/otndocs/jcp/7224-javabeans-1.01-fr-spec-oth-JSpec/) which is only for `boolean` not `Boolean`.

As suggested in the ticket, Freemarker should probably improve its documentation that `isX()` works with `boolean, not with `Boolean`.

mapeders:
Thanks @chrisrueger — fair point on 8.3.2, the current behavior is strictly
  spec-compliant.                                                               
                                                                                
  That said, the FreeMarker manual already says model.foo invokes getFoo() or   
  isFoo() without qualifying the return type, and the change here is gated      
  behind incompatible_improvements >= 2.3.35 (mirroring the existing Java 8
  default-method supplement path in ClassIntrospector) — so existing configs see
   no change.               

  Happy either way though: if the committers prefer a docs-only fix, I'll gladly
  close this and open a documentation PR instead.

ddekany:
While this is not a bug, it's certainly quite annyoing and people keep running into it. But I will have to review the implementation to see if there's any cost involved that everyone has to pay, etc.

